PR #6, opened 2025-07-03
AGPF-127: Turn validations for `is.null(.alt_df_full)` into a function

rmao6:
https://jira.jnj.com/browse/AGPF-127


munoztd0:
LGTM

PR #19, opened 2025-07-07
Dev varying dp

iaugusty:


iaugusty:
other approach is under development

PR #21, opened 2025-07-08
only exporting 4 h_* functions

eanokian:
- removed from NAMESPACE the exported h_* functions except 4
- suppressed the messages in the unit tests so now devtools::test() does not show any messages (only messages are hidden)
- renamed test_markup_unicodify.R to test-markup_unicodify.R, so it is now included in the unit tests
- fixed a type in test-rbmi.R ("requireNamesapce")


munoztd0:
LGTM thx!

Issue #33, opened 2025-07-21
fix typos in colwidths.R functions ttype_wrap_vec() and constrict_lbl_lns()


eanokian:
function **ttype_wrap_vec()** is only being called in **check_wrap_nobreak()**
function **check_wrap_nobreak()** is not being used, but it's being exported, so we decide to keep both functions.

The following fix was applied to **ttype_wrap_vec**:
<img width="629" height="184" alt="Image" src="https://github.com/user-attachments/assets/a3f488fd-f6cf-438a-aa76-bafb13537f6a" />

The following fix was applied to **constrict_lbl_lns**:
<img width="699" height="165" alt="Image" src="https://github.com/user-attachments/assets/ffebbb91-09d2-4445-be2a-2267a7551273" />

eanokian:
also potentially remove these functions which seems to be unused:

in R/colwidths.R, function **find_free_colspc()**

eanokian:
done so far:

- **find_free_colspc**: not used, not exported -> removed, left unit tests commented out
- **smart_colwidths_1page**: not used, not exported -> removed, no unit tests
- **combine_nonblank**: not used, not exported -> removed, no unit tests
- **rm_blank_levels**: not used, not exported -> removed, no unit tests
- **ttype_wrap_vec**: used only by check_wrap_nobreak, not exported -> kept it and unit tests
- **check_wrap_nobreak**: not used, exported -> kept it and unit tests

PR #34, opened 2025-07-24
152 increase coverage

eanokian:
1. added more unit tests to increase coverage
2. wrapped function "make_rbmi_cluster" with nocov
3. fixed function argument partial matching
4. the following fix was applied to **constrict_lbl_lns**:
Image

5. the following fix was applied to **ttype_wrap_vec**:
Image


eanokian:
addresses issue #33

munoztd0 (approved):
LGTM

PR #51, opened 2025-08-11
Moving out from old CI/CD to new one

munoztd0:
Only changes to github actions that need to be integrated in main. No actual "code" changes


munoztd0:
This error is expected this is actually the purpose of this PR
Inspection 🛠 / Check URLs 🌐 / URL Checker 🔎 (pull_request)Failing after 43s
After this merge then the unit-test-reporter will be created and this URL will exist

jkovach2 (approved):
No comment.

PR #112, opened 2025-11-21
Release rbmi suggests

munoztd0:
# Pull Request

I moved `rbmi` to `Suggests` and added the R-Universe repo. R CMD check still shows an error:

```
Package suggested but not available: 'rbmi'
```

However, it does detect availability via `Additional_repositories`:

```
rbmi   yes   https://insightsengineering.r-universe.dev/
```

So there’s a chance we might be able to work something out, even though CRAN flags it because it’s archived.


munoztd0:
> @munoztd0 I now added more rbmi functions tests, with this we should have sufficient coverage for them.

With those and my final fixes I think we are good @gmbecker ?

PR #114, opened 2025-11-21
a_maxlev fun update: `denom_df` argument replaced by `.alt_df_full`.

wwojciech:
# Pull Request

Replaced `denom_df` with `.alt_df_full` in `a_maxlev()`.

## Checks


munoztd0:
I mean if it needs to be merged quickly I don't see why but I will need you do either screenshot here a devtools::check() result that pass all tests or send it to me via PM.

PR #149, opened 2026-01-13
this commit addresses #65, by adding a preemptive pruner approach: if…

### LINKED ISSUE #65: count_pruner efficiency
See Issue - [count_pruner generated function calls subset_cols and col_paths too much · Issue #3 · johnsonandjohnson/junco](https://github.com/johnsonandjohnson/junco/issues/3)

Also review AGPF-142 from JIRA (https://jira.jnj.com/browse/AGPF-142).

eanokian:
# Pull Request

Fixes #65

## Checks

### DIFF
--- a/R/pruning_functions.R
+++ b/R/pruning_functions.R
@@ -123,14 +123,28 @@ count_pruner <- function(count = 0, cat_include = NULL, cat_exclude = NULL, cols
   function(tt) {
     # Do not ever prune the following rows.  a row that should be kept in the table will get the value of FALSE
 
+    # preemptive pruning
+    # if current row is a "ContentRow" that contains all zeros,
+    # return TRUE (i.e. prune current row and all its children)
+    preemptive_pruning_possible <- FALSE
+    if (methods::is(tt, "TableTree")) {
+      children_types <- make_row_df(tt) |> pull(node_class)
+      if (length(which(children_types == "ContentRow")) == 1 &&
+            children_types[1] == "ContentRow") {
+        preemptive_pruning_possible <- TRUE
+      }
+    }
+
+
     if ( # nolint start
-      !methods::is(tt, "TableRow") ||
+      !preemptive_pruning_possible && (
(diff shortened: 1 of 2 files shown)


gmbecker:
Not sure if this is still a live PR, if not please close it @eanokian. If it is, please handle merge requests and add tests to ensure behavior stays the same

eanokian:
@gmbecker  it is a live PR, it's a topic I investigated in Jan but I'd like to discuss with you. We are in a point where we need to decide whether to incorporate this minor fix to dev or discard it all along.
please look at the issue https://github.com/johnsonandjohnson/junco/issues/65 for context.
Not urgent though and can be discussed next week.

R/pruning_functions.R
```diff
@@ -123,14 +123,28 @@ count_pruner <- function(count = 0, cat_include = NULL, cat_exclude = NULL, cols
   function(tt) {
     # Do not ever prune the following rows.  a row that should be kept in the table will get the value of FALSE
 
+    # preemptive pruning
+    # if current row is a "ContentRow" that contains all zeros,
+    # return TRUE (i.e. prune current row and all its children)
+    preemptive_pruning_possible <- FALSE
```
gmbecker, line 129: I'm not super convinced we need to do this given the minimal time gain you cite in #65 but if we do, it should be cached like `col_paths` is above. There's no reason to run this logic 200 times.

eanokian: mhh I'm not sure to follow you.
This variable should be specific to each iteration/recursive function call. It may happen that it's TRUE for particular slices of the original table and FALSE for other ones.
If we can accomplish that logic with the caching strategy, then it makes sense to cache it, but I thought caching happens only once and it's global across all iterations or recursive calls, but I might have misunderstood it.
In any case, I agree with you that the gain is minimal (if any) and for me we can disregard this PR altogether.

eanokian:
closing this issue as most likely it won't be merged into dev. We can reopen this conversation in the future if needed.

PR #167, opened 2026-01-22
Hotfix: Argument "label_width_ins" in tt_to_tlgrtf() not passed recursively #166

### LINKED ISSUE #166: [Bug]: bugfix argument "label_width_ins" in `tt_to_tlgrtf()` not passed recursively
### What happened?

argument "label_width_ins" in `tt_to_tlgrtf()` needs to be passed recursively
right now it's being reset to the default (= 2)

### sessionInfo()

```R

```

### Relevant log output

```R

```

eanokian:
# Pull Request

Fixes #166 

## Checks

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: junco
 Title: Create Common Tables and Listings Used in Clinical Trials
-Version: 0.1.3
+Version: 0.1.3.9
 Date: 2026-01-12
 Authors@R: c(
     person("Gabriel", "Becker", , "gabembecker@gmail.com", role = c("cre", "aut"),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.3.9] - Unreleased
+
+### Fixed
+- Fixed `tt_to_tlgrtf()` argument `label_width_ins` which was not applying the change in the row label column width (#166).
+
 ## [0.1.3] - 2026-01-12
 
 ### Changed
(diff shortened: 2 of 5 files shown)


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: junco
 Title: Create Common Tables and Listings Used in Clinical Trials
-Version: 0.1.3
+Version: 0.1.3.9
```
jkovach2, line 3: @eanokian , why 0.1.3.9? This is just for my own understanding.

jkovach2:
@eanokian  do we need to update changelog.md? I don't see that updated in the "Files changed". Also, do we need to incorporate a test to ensure that changing the label_width_ins argument value actually changes the column width for the label column?

DESCRIPTION
munoztd0: It’s basically a hotfix-version marker. After a CRAN release like 0.1.3, we often use 0.1.3.9 or 0.1.3.9000 to signal “we’re past the last release but not at the next one yet.” It’s a clear way to show ongoing development without bumping the official version.

jkovach2:
@munoztd0 @eanokian @iaugusty @gmbecker is there a reason why we don't export tt_to_tbldf? A user noticed this in SPACE when they were attempting to create a QC data frame. I pointed them to use the triple colon to access the function, howevr, it got me thinking, was it intentional for us not to export this function, knowing that users actively call it?

eanokian:
> @eanokian do we need to update changelog.md? I don't see that updated in the "Files changed". Also, do we need to incorporate a test to ensure that changing the label_width_ins argument value actually changes the column width for the label column?

yes, I just pushed the updated changelog.md.

I also added unit tests that test different values of argument `label_width_ins`: 2 (the default), 5 and 0.5. You can see them in this commit:
https://github.com/johnsonandjohnson/junco/pull/167/commits/9c256d0d348663565a0c74a489c98fe11052b3b2

munoztd0 (approved):
Nice work around for testing the correct width ! LGTM

PR #203, opened 2026-02-13
Export helpers for LS means tabulation

danielinteractive:
This exports helpers for wide tabulation of LS means results which were only used internally before in the layout generating function. I have not added this to the NEWS as it is a very minor update.


munoztd0:
updated changelog accordingly

munoztd0 (approved):

PR #219, opened 2026-02-25
218 docx vertical pagination reduce first col indent

### LINKED ISSUE #218: docx vertical pagination + reduce first col indentation
In some edge cases, when we have vertical pagination and sub-parts, in function `tt_to_flextable_j()`, the Boolean argument 'reduce_first_col_indentation' is not calculated correctly, leading to row labels (first column in the body) having all their indentation reduced by 1 when they shouldn't.

Besides, function `insert_keepNext_vertical_pagination()` is inserting too many page breaks, leading the output docx to have too many pages, each having just a few rows.

eanokian:
# Pull Request

Fixes #218 

In some edge cases, when we have vertical pagination and sub-parts, in function `tt_to_flextable_j()`, the Boolean argument 'reduce_first_col_indentation' was not calculated correctly, leading to row labels (first column in the body) having all their indentation reduced by 1 when they shouldn't.

Besides, function `insert_keepNext_vertical_pagination()` was inserting too many page breaks, leading the output docx having too many pages, each with just a few rows.

This PR addresses both issues, by improved calculation of `reduce_first_col_indentation`, and improved function `insert_keepNext_vertical_pagination()`, which now calculates the page breaks in the same way as `tt_to_tlgrtf()`.

## Checks


munoztd0 (approved):
lgtm !

PR #227, opened 2026-03-10
117 ancova combined column revised

### LINKED ISSUE #117: [Bug]: afuns that call s_ancova_j in a layout with combined groups result in error
### What happened?

A bug happened!
afuns that call s_ancova_j or s_summarize_ancova_j in a layout with combined groups result in error. 
The following 2 afuns are impacted: a_summarize_aval_chg_diff_j and a_summarize_ex_j

### sessionInfo()

```R

```

### Relevant log output

```R

```

iaugusty:
# Pull Request

Fixes #117

## Checks


iaugusty:
@danielinteractive : could you take a look at the proposed approach for dealing with estimations from ancova model in combined column? I'm using a weighted approach where the proposed weights are independent from the existing weights_emmeans parameter. Your input would be highly appreciated, before going into discussion with stats.
No default has been set (on purpose) and the user would have to make their choice regarding weights_combo.

munoztd0:
All good as for our templates at least -> https://github.com/munoztd0/scda.test/actions/runs/22907556675/job/66470541626?pr=4

danielinteractive:
Thanks @iaugusty , I think it will be best to first discuss with you to understand more the context and idea

iaugusty:
@danielinteractive : Stats team has given confirmation that proposed approaches make sense, can you review the PR?

tests/testthat/test-summarize_ancova_combined.R
```diff
@@ -0,0 +1,749 @@
...
+
+    sum_4 <- list(est_red = derived$combined$red, est_blue = derived$combined$blue)
+  }
+  ret <- list(tbl1 = sum_1, tbl2 = sum_2, tbl2_comb = sum_2b, derived = sum_3, comb_model = sum_4)
+
+  ret
+}
+
+
+expect_any_difference <- function(object, expected, ...) {
```
danielinteractive: This general function would be better placed in a `helper_*.R` file

tests/testthat/test-summarize_ancova_combined.R
```diff
@@ -0,0 +1,749 @@
...
+iris_plus2 <- iris_plus |>
+  group_by(Species) |>
+  mutate(id = row_number()) |>
+  filter(Species == "setosa" & id < 45 |
+    Species == "versicolor" & id < 30 |
+    Species == "virginica" & id < 50) |>
+  ungroup()
+
+
+test_that("summarize_ancova cannot deal with a combined column", {
```
danielinteractive: maybe also mention that this function is from the tern package - so this is motivating the development of the junco function

tests/testthat/test-summarize_ancova_combined.R
```diff
@@ -0,0 +1,749 @@
...
+  expect_false(
+    isTRUE(comp),
+    info = "Objects are identical; expected at least one difference"
+  )
+}
+
+
+#' actual tests starts here
+
+test_that("s_ancova_j works as expected in combined column for model with interaction 3 sets of weights_combo", {
```
danielinteractive: the description here mentions `s_ancova_j` but this is called only indirectly through the wrapper functions defined above, which makes the tests hard to read. I would recommend to 
1) shorten each test, and distribute test code to multiple tests (i.e. `test_that()` calls) with specific description
2) better use direct calls to `s_ancova_j` if possible (or `a_summarize_ancova_j` and adjust test description accordingly), even if that leads to some duplicated code between tests
3) try to reduce or avoid wrapper functions defined above

tests/testthat/test-summarize_ancova.R
```diff
@@ -61,7 +61,7 @@ test_that("s_ancova_j works as expected", {
   checkmate::expect_names(
     names(result),
     identical.to = c(
-      "n",
+      "n_fit",
```
danielinteractive, line 90: this should also be changed below

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added argument 'validate' to `export_TLG_as_docx()` and `tt_to_flextable_j()` (#213)
 - added "watermark" argument in the docx exporter for Figures (#181)
 - Export helpers for LS means tabulation: `lsmeans_wide_cfun`, `lsmeans_wide_first_split_fun_fct`, `lsmeans_wide_second_split_fun_fct`
+- Added option to perform `a_summarize_ancova_j` in a layout with a combined column facet (#117)
```
danielinteractive: also mention the slight change in return value for `s_ancova_j`, now we have `n_fit` instead of `n` returned

R/junco_utils_default_stats_formats_labels.R
```diff
@@ -195,14 +195,44 @@ junco_default_stats <- list(
   relative_risk = c("rel_risk_ci", "pval"),
   summarize_ancova_j = c(
     "n",
```
danielinteractive, line 195: do we still have `n` now as statistic? just double checking

R/summarize_ancova.R
```diff
@@ -42,10 +42,247 @@ h_ancova <- function(
...
+                                .in_ref_col) {
+  # Ensure that there is only one element in sum_level.
+  checkmate::assert_scalar(sum_level)
+
+  sum_fit <- summary(
+    emmeans_fit,
+    level = conf_level
+  )
+
+  sum_fit_level <- sum_fit[sum_fit[[arm]] == sum_level, ]
```
danielinteractive: better be explicit here whether we want to drop dimensions or not using the `drop` argument

R/summarize_ancova.R
```diff
@@ -42,10 +42,247 @@ h_ancova <- function(
   )
 }
```
danielinteractive, line 44: please add documentation chunk

R/summarize_ancova.R
```diff
@@ -42,10 +42,247 @@ h_ancova <- function(
...
+    }
+  }
+  ret <- list(
+    sum_fit_level = sum_fit_level,
+    sum_contrasts_level = sum_contrasts_level,
+    y = y
+  )
+  ret
+}
+
```
danielinteractive, line 159: please add documentation chunk

R/summarize_ancova.R
```diff
@@ -42,10 +42,247 @@ h_ancova <- function(
...
+make_contrast_weights <- function(weight_mode = c("equal", "proportional"),
+                                  normalize = TRUE,
+                                  sum_fit,
```
danielinteractive: better move arguments without a default value to the start of the argument list in the function definition, and arguments with a default value at the end of it

R/summarize_ancova.R
```diff
@@ -42,10 +42,247 @@ h_ancova <- function(
...
+  if (normalize) {
+    w <- w / sum(w)
+  }
+
+  # update the weights for sum_levels with the above values
+  w_all[names(w)] <- w[names(w)]
+  w_all
+}
+
+h_ancova_est_combined <- function(emmeans_fit,
```
danielinteractive: please add documentation chunk

R/summarize_ancova.R
```diff
@@ -42,10 +42,247 @@ h_ancova <- function(
...
+  }
+
+  contr <- emmeans::contrast(emmeans_fit, list(
+    setNames(list(w_start[["weights1"]]), "Estimate contrast"),
+    setNames(list(w_start[["weights2"]]), "Estimate contrast vs ref")
+  ),
+  # Derive confidence intervals, t-tests and p-values.
+  infer = TRUE,
+  # Do not adjust the p-values for multiplicity.
+  adjust = "none"
```
danielinteractive: check indentation

R/summarize_ancova.R
```diff
@@ -400,7 +605,18 @@ a_summarize_ancova_j <- function(
     ref_path,
     .spl_context,
     ...,
-    .stats = NULL,
+    .stats =
+      c(
+        "n",
```
danielinteractive: and/or `n_fit`?

danielinteractive (commented):
Thanks @iaugusty, please see comments below. 
I wonder if in addition a more high level description of the functionality, along side a small example, in a vignette would be helpful for the users and long term maintenance.

R/summarize_ancova.R
iaugusty: n_fit is only relevant for models with interaction, which is not default, suggest not to add n_fit as default statistic

tests/testthat/test-summarize_ancova_combined.R
iaugusty: can we connect to discuss workable approach here, otherwise test scripts were too long/complicated

R/junco_utils_default_stats_formats_labels.R
iaugusty: yes, n is from descriptives (s_summary(x) and n_fit is from ancova part (s_ancova_j)

R/summarize_ancova.R
iaugusty: I'll add

danielinteractive (approved):
looks much better, thanks @iaugusty !

PR #231, opened 2026-03-11
Hotfix: boxplots whose statistics follow SAS percentile definition (`PCTLDEF = 5`)

munoztd0:
# Pull Request

Fixes issue with quantile default type = 7 in geom_boxplot()

## Checks


R/geom_j.R
```diff
@@ -0,0 +1,92 @@
...
+  required_aes = c("x", "y"),
+  dropped_aes = c("x", "y", "weight"),
+  compute_group = function(data, scales, width = NULL, na.rm = FALSE, coef = 1.5) {
+    vec <- data$y
+    if (length(vec) == 0) {
+      return(data.frame())
+    }
+
+    # SAS PCTLDEF=5 (Type 2 Quantiles)
+    qs <- quantile(vec, probs = c(0, 0.25, 0.5, 0.75, 1), type = 2, na.rm = na.rm)
```
gmbecker: how difficult would it be to make quantile_type an argument rather than hardcoded here?

R/geom_j.R
```diff
@@ -0,0 +1,92 @@
+#' SAS-style boxplot statistics
```
gmbecker: I think we can mention SAS but I think this goes too far as drafted. I think describe it as using different a different quantile definition (or definitions, see comment below) with one sentence that the default we use is the one used by SAS

R/geom_j.R
gmbecker: judging by a quick reading of https://ggplot2.tidyverse.org/articles/extending-ggplot2.html it looks like it should be easy to do

gmbecker (commented):
overall looks reasonable, if we're doing this I think its worth doing generally (with the default being what we want) see comments.

R/geom_j.R
munoztd0: Definitely not difficult, I was just a bit too focus on the "SAS-replacement" that I didn't even thought about opening it to others quantile calculations but now it's resolved

R/geom_j.R
munoztd0: Adressed that

munoztd0:
Created an issue in https://github.com/tidyverse/ggplot2/issues/6819

munoztd0:
The check errors seem to be weird false positives from export_docx.R since I didn't touch anything form there.. @eanokian can you confirm ?

eanokian (approved):
lgtm, nice work!

eanokian:
> The check errors seem to be weird false positives from export_docx.R since I didn't touch anything form there.. @eanokian can you confirm ?

yes, we still have the old unit tests in this branch, should we update them from another branch, like 'dev'? as they are now, they are doing full snapshot test on XML content of docx files.

munoztd0:
> > The check errors seem to be weird false positives from export_docx.R since I didn't touch anything form there.. @eanokian can you confirm ?
> 
> yes, we still have the old unit tests in this branch, should we update them from another branch, like 'dev'? as they are now, they are doing full snapshot test on XML content of docx files.

We just need to pass these two hotfixes, so my choice would be to accept the outdated snapshots for now, since we are note shipping the new docx code anyhow.

PR #234, opened 2026-03-13
Added Hotfix pipeline

munoztd0:
We needed a way to test that our hotfixes where properly tested.

Thus I added it to a folder named **dev** (which is obviously ignored from Rbuilds)


munoztd0:
@gmbecker added a new **pipeline check** -> https://github.com/johnsonandjohnson/junco/actions/runs/23058198672/job/66977367809?pr=234

Not my most pretty pipeline but working outside of it's encapsulation is quite messy.
I works well since it's produce `conflicts` and doesn't create a separate environement with `estthat::test_local()`

PS: the failed test are still because of old docx_exporter erroneous snapshots (fixed in #231 )

PR #240, opened 2026-03-17
update: remove private data source

### LINKED ISSUE #239: tests: sas_comparison folder
What is this folder and why is it here ?

https://github.com/johnsonandjohnson/junco/blob/46c36d1e38f621e665877d6c859a3a36409b81b5/tests/testthat/sas_comparison/

@danielinteractive ?

munoztd0:
# Pull Request

Fixes #239 

@danielinteractive if you need you can put the data example2.csv in the /dev folder but we should not put JnJ private paths in our gtihub public code


danielinteractive (approved):
Agree, thanks - I guess we should have cleaned this up before open sourcing.

PR #255, opened 2026-04-01
Add multiplicity adjustment option to `fit_mmrm_j()`

danielinteractive:
Implement functionality to adjust p-values and confidence intervals for least square means contrasts using Dunnett's method and step-down Dunnett adjustment. Update tests and documentation accordingly.


munoztd0 (changes requested):
Well I understand hat you run a air format on the whole package, which is nice but since we didn't do that a the beginning it created a bunch of changes .. so could you just not push them in your PR (it even created lint issues I believe.)
We try to keep the PR has lean as possible.

munoztd0 (commented):
update docs and export h_get_mult_adj_estimates

@danielinteractive I added export to h_get_mult_adj_estimates() since you don't use it internally I believe you need it for templates and thus needs to be exported right ?

munoztd0 (approved):
If you validate the export

danielinteractive:
Thanks @munoztd0 , `get_mmrm_lsmeans()` is using this internal helper, which should not be exported - I modified accordingly.

munoztd0:
> Thanks @munoztd0 , `get_mmrm_lsmeans()` is using this internal helper, which should not be exported - I modified accordingly.

🤦 I somehow missed it in my search .. sorry

danielinteractive:
no prob, thanks @munoztd0 !

PR #258, opened 2026-04-10
introduce safe_t_test with tryCatch to prevent problems with almost c…

### LINKED ISSUE #257: [Bug]: s_summarize_desc_j fails with almost constant data - t.test
### What happened?

A bug happened!
``` r
library(junco)
#> Loading required package: formatters
#> 
#> Attaching package: 'formatters'
#> The following object is masked from 'package:base':
#> 
#>     %||%
#> Loading required package: rtables
#> Loading required package: magrittr
#> 
#> Attaching package: 'rtables'
#> The following object is masked from 'package:utils':
#> 
#>     str
#> Registered S3 method overwritten by 'tern':
#>   method   from 
#>   tidy.glm broom

df <- data.frame(
  "AVAL" = c(1.709999999999999964473, 1.710000000000000186517, 1.710000000000000186517, 1.710000000000000186517, 1.710000000000000186517)
)
.ref_group <- df
.in_ref_col <- FALSE

stats <- junco:::s_summarize_desc_j(df, "AVAL", .ref_group, .in_ref_col)
#> Error in `t.test.default()`:
#> ! data are essentially constant
```

### sessionInfo()

```R

```

### Relevant log output

```R

```

iaugusty:
…onstant data

# Pull Request

Fixes #257

## Checks

- [ x] (Have you updated the changelog.md ?)


iaugusty:
would this sufficient as bug fix?

R/a_summarize_aval_chg_diff.R
```diff
@@ -12,27 +46,17 @@ s_summarize_desc_j <- function(df, .var, .ref_group, .in_ref_col, control = cont
     x1 <- x1[!is.na(x1)]
     x2 <- x2[!is.na(x2)]
 
-    if ((length(x1) > 1 && length(x2) > 1) && !(stats::var(x1) == 0 && stats::var(x2) == 0)) {
-      ttest_stat <- stats::t.test(x1, x2, conf.level = control$conf_level)
+    ttest_stat <- safe_t_test(x1, x2, conf.level = control$conf_level)
```
munoztd0, line 50: I see that you added safe_t_test because of the "constant" bug, but now you dropped the if/else block, and I don't get why because that would drop the fallback behavior for groups with $n < 2$. Without that block, an empty group triggers the tryCatch which runs mean() on an empty vector, resulting in NaN. I think we should keep it wrapped in the if/else block so we keep both behaviors like so:

```
  if (length(x1) > 1 && length(x2) > 1) {
      ttest_stat <- safe_t_test(x1, x2, conf.level = control$conf_level)
      
      stat <- ttest_stat[c("estimate", "conf.int")]
      stat$diff <- stat$estimate[1] - stat$estimate[2]
      stat <- c(stat$diff, stat$conf.int)
    } else {
      y2b <- s_summary(.ref_group[[.var]])
      diff <- y1[["mean"]] - y2b[["mean"]]
      stat <- c(diff, NA_real_, NA_real_)
    }
```

munoztd0 (changes requested):

R/a_summarize_aval_chg_diff.R
iaugusty: Can't we better address this inside safe_t_test by converting NaN into NA? 

mean(numeric(0)) = NaN
whereas s_summary(numeric(0))[["mean"]] = NA

so when we would replace NaN inside safe_t_test function, there is no longer a need for 2 separate blocks?
````
estimate[is.nan(estimate)] <- NA_real_
````

munoztd0: That works for me too !

munoztd0 (approved):
Good for me now

PR #271, opened 2026-04-15
CRAN ready: Adding multi-comparator functionality  lyt fun

gmbecker:
Not intended for immediate merge, but for team to look through and review approach, with merge after discussions.

## Checks


R/risk_diff_col_struct.R
```diff
@@ -0,0 +1,688 @@
...
+    splvar <- comp_vars[1]
+    more_comp_vars <- comp_vars[-1]
+    all_lvls <- levels(fulldf[[splvar]])
+
+    if (is.null(comp_map)) {
+      comp_map <- make_dflt_comp_map(fulldf, splvar, ref_lvls, combo_map, comp_vars)
+    }
+
+    lvls_to_keep <- levels_from_comp_map(comp_map, fulldf, ref_lvls, comp_vars, combo_map)
+    restrict_facets(lvls_to_keep, op = "keep")(ret, spl, fulldf)
```
iaugusty, line 364: @gmbecker where is this function defined? I get 
`Error applying custom split function: could not find function "restrict_facets"`
when running tests

munoztd0: Same here

R/risk_diff_col_struct.R
```diff
@@ -0,0 +1,688 @@
...
+
+  main_splfun <- make_split_fun(pre = .main_pre, post = main_post)
+
+  rr_splfun <- make_multicomp_splfun(
+    comp_vars,
+    ref_paths,
+    comp_level_map = comp_map,
+    combo_levels_map = combo_map_df,
+    .pre = .rr_pre,
+    post = .rr_post
```
iaugusty: `.post = .rr_post`

R/risk_diff_col_struct.R
```diff
@@ -0,0 +1,688 @@
...
+      colspan_trt_map <- add_combo_levs_to_trtmap(colspan_trt_map, combo_map_df)
+    } else if (any(!combo_found)) {
+      warning("some combination levels defined in combo_map_df do not appear in colspan_trt_map")
+    }
+  } else {
+    main_post <- list()
+  }
+
+  ctrl_span <- unique(colspan_trt_map[[spanvar]])[2] ## assume second is non-active
+  ctrl_vals <- colspan_trt_map[colspan_trt_map[[spanvar]] == ctrl_span, trtvar]
```
iaugusty: shouldn't it take into account that less control groups can be requested via comp_map? Eg only against Pbo?

R/risk_diff_col_struct.R
gmbecker: Ah, sorry, I forgot to update the description file. Its in rtables in aPR that I had thought was merged in but it wasn't. should be soon: https://github.com/insightsengineering/rtables/pull/1081

R/risk_diff_col_struct.R
gmbecker: Technically, I *think* this is better interpreted as "possible control groups", if a comp_map is specified, that will be used when constructing the actual risk difference columns. 

I agree we need a test to confirm this though, and if it is causing levels to show up where they shouldn't we'll need to fix it.

gmbecker:
@iaugusty  @munoztd0  @jkovach2  @eanokian reworked as discussed and ready for another look through

dev/tests/double_risk_test.R
munoztd0, line 1: we shoudl remove this, I jsut put here so we could actually run the tests

R/split_functions.R
```diff
@@ -386,3 +386,33 @@ do_exclude_split <- function(exclude_levels, .spl_context) {
...
+
+
+
+
+insert_subset_exprs <- function(partinfo, spl, comp_path = NULL) {
+  rvs <- rawvalues(partinfo$values)
+  names(rvs) <- names(partinfo$values)
+  exprs <- lapply(rvs, function(rvi) make_subset_expr(spl, rvi))
+  newvals <- mapply(function(val, expr) {
+    exvals <- utils::getFromNamespace("splv_extra", "rtables")(val)
```
munoztd0, line 398: wrapped in getFromNamespace because non exported rtables function ?

DESCRIPTION
munoztd0, line 1: modified this file to reflect our version dependency state

NAMESPACE
munoztd0, line 1: ran devtools::document() are thos right exports wise?

munoztd0 (approved):
After addressing my comments and my last commits

R/risk_diff_col_struct.R
```diff
@@ -0,0 +1,691 @@
...
+  names(out) <- names(orig_ret)
+  out
+}
+
+get_comp_label <- function(comp_level, orig_ret, combo_map) {
+  if (!is.null(combo_map) && comp_level %in% combo_map$valname) {
+    ret <- unique(combo_map$label[combo_map$valname == comp_level])
+  } else {
+    ind <- which(comp_level == names(orig_ret$values))
+    if (length(ind) != 1) browser()
```
munoztd0: @gmbecker I think you forgot this browser() !
Hope this is not the version you put it CRAN

munoztd0:
```suggestion
    if (length(ind) != 1)
```

PR #277, opened 2026-04-21
Add `subgroup` variable option for `fit_mmrm_j()` and downstream tabulation

### LINKED ISSUE #274: Add subgroup option to MMRM stack
We need a way to have a subgroup factor variable enter the MMRM stack via the `vars` argument of `fit_mmrm_j` and then pass that downstream. This shall enable a subgroup specific treatment effect estimation, however with joint parameters for the other covariate coefficients as well as the variance parameters.

- [x] subgroup * visit * treatment needs to be included when `subgroup` element of `vars` is a factor and not empty (default)
- [x] modify `get_mmrm_lsmeans` which is called in `fit_mmrm_j`
  - [x] it should also work with averages of visits
  - [x] it should also work with multiplicity adjustment
- [x] and then also need to adapt `tidy.tern_model`

danielinteractive:
closes #274


munoztd0:
@danielinteractive can you check the snapshot fails ?

munoztd0:
Also the copilot PR review is pretty neat and thorough, so I don't know if you want me to help with it or would you check the suggestions ?

danielinteractive:
Thanks @munoztd0 , I will check both now

danielinteractive:
@munoztd0 ok hopefully all ok now from automatic test side, ready for manual review from your side

munoztd0 (approved):
lgtm thanks for going through the whole process

PR #321, opened 2026-05-09
Fixes #312: Add stats to `s_diff_mean_ci()`, and rename it to `s_diff_means()`.

### LINKED ISSUE #312: [Feature Request]: Add more stats to `s_diff_mean_ci()`
### Feature description

1. Add more statistics to `s_diff_mean_ci()` function:

`diff_means_n1` - sample Size of Group 1
`diff_means_n2` - sample Size of Group 2
`diff_means_est` - point estimator of the difference in population means
`diff_means_se` - standard error of the estimator of the difference in population means
`diff_means_est_se` -  point estimator and standard error of the estimator of the difference in population means
`diff_means_ci` - confidence interval for the difference in population means
`diff_mean_est_ci` - point estimator and the confidence interval for the difference in population means

2. Rename `s_diff_mean_ci()` ->  `s_diff_mean()`.

wwojciech:
Fixes #312 

## Checks


R/safe_t_test.r
```diff
@@ -24,16 +24,17 @@
 #' field containing the error message.
 #'
 #' @importFrom stats t.test
-#' @export
 #'
 #' @examples
 #' # Standard usage
 #' t.test(1:10, 11:20)
-#' safe_t_test(1:10, 11:20)
+#' junco:::safe_t_test(1:10, 11:20)
```
munoztd0: This unexported symbol will cause R CMD check to fail on CRAN.
We should either export it if it's meant for the public API, or remove it from this example.

munoztd0 (changes requested):
So overall, I'm ok with it but as we said this would be a breaking change for the TLG catalog, so 

1) fix the unexported safe_t_test example
2) create a ticket with all the affected templates (just a screenshot of a VScode search window is ok ) and link it to this PR

R/safe_t_test.r
wwojciech: Will it also fail when I surround it with `notrun{}` ?

wwojciech:
> So overall, I'm ok with it but as we said this would be a breaking change for the TLG catalog, so
> 
> 1. fix the unexported safe_t_test example
> 2. create a ticket with all the affected templates (just a screenshot of a VScode search window is ok ) and link it to this PR

Thanks @munoztd0 - so actually, only one table will be affected (jjcs-core TSFVIT01). I will create a ticket for that.

R/safe_t_test.r
munoztd0: It **will** pass but in this case it useless to add an example that **will** fail , are we demonstrating is , should the user be capable to copy paste run it: yes -> exported / no -> not in user-facing example, maybe in unit test for devs.

Make sense ?

wwojciech: I consider this function internal (not exported), and the examples are intended to help users (i.e., package developers / internal users) understand its purpose and the design rationale behind it.

For that reason, I would prefer to keep the examples as they are, since they illustrate meaningful behavior that is otherwise not obvious.

However, I agree they should not fail CRAN checks or break example execution. If needed, we can wrap them in \dontrun{} (or \donttest{} depending on intent) to make it explicit that they are illustrative and not meant to be executed during checks or by end users.

munoztd0: I appreciate wanting to keep the examples for internal developers, but as the repo owner I won't approve merging this using \dontrun{}. Since the function is not exported, the standard best practice is to add #' @noRd to the roxygen block. This leaves all your examples untouched in the source code for our developers to read, but prevents the .Rd file from generating, which naturally bypasses the CRAN checks.


[Cran Policies](https://contributor.r-project.org/cran-cookbook/general_issues.html)

image

Which this case do not belong

wwojciech: The CRAN policy says that these are the "most common" examples; this does not mean that other examples are unacceptable.

In summary, I do not see any real benefits to excluding this from the examples section, while I see at least two good reasons to keep it there. One has already been mentioned. The second is that not all exported functions are documented sufficiently well. Therefore, sometimes a regular user of the package (not a developer) may need to inspect the code to quickly understand what `safe_t_test()` does. In such cases, these examples are very helpful.

munoztd0: OK I don't think you understand my point, it's not about "benefits" it is just simply that CRAN _will_ give a NOTE and _will_ not be happy about it and we will fail automatic review processes and have to make a case for it. I have done that on first release and won't do that now . You have to also think about maintainers and CRAN reviewers in your equation. That is it - EOD. No ::: colon in package space (including examples, and including example in notrun{}).

That being said I agree with most of you say so we should just document quickly add a test and export it, if you don't want or have time to do it, I can take over happily. 

But regarding to the ::: , it is just NO.

wwojciech: Indeed, I was not aware about the NOTE from CRAN. Ok, then.

munoztd0 (approved):
All good now on my end we can merge

wwojciech:
> All good now on my end we can merge

Perfect, than I think we can merge, because I will anyway discuss with Ilse the overall design again. (we are in the middle of discussing it already).

PR #330, opened 2026-05-13
Solves 328: Expand, refactor and rename insert_blank_line() to insert_line().

### LINKED ISSUE #328: Extend `insert_blank_line()` so that it can create a custom label
### Feature description

Extend `insert_blank_line()` so that it can create a custom label. It should accept an `afun` argument: an analysis function that generates a label and returns a single `CellValue` S3 object, that is, `rtables::rcell(NULL, label = custom_label)`, where the `custom_label` is created by this customer `afun` function. 

This would allow dynamic row labels to be created in a convenient and scalable way, which is needed, for example, for the jjcs-core `TSFVIT01` table.

wwojciech:
Fixes #328 

## CAUTION!

This changes requires update of several TLG scripts, since `insert_line()` is exported function. This is a massive update but very simple: `insert_blank_line()` must be replaced with `insert_line()`. No any other changes are required.
Issue #331 is created for this task.

## Checks


wwojciech:
Hey @munoztd0 - how we could proceed with this? Thanks!

munoztd0:
@wwojciech , I do not know, this quite a bit of a breaking change, we should talk about this on the team meeting

wwojciech:
After discussion with the team, we decided not to generalize `insert_blank_line()` because it has a very specific and limited use case.

PR #348, opened 2026-05-26
Modified and renamed a_summary_j() to a_summary_subset(), modified c_…

### LINKED ISSUE #341: Refactor `a_summary_j()`, `c_summary_subset_label()`.
Tasks:

- [x] Rename `a_summary_j()` to `a_summary_subset()`. 
- [x] `a_summary_subset()` should simply filter rows and then invoke `tern::a_summary(x, ...)` with `...` for the additional optional arguments (e.g. formatting).
 Remove `subset_expr` to `filter_expr`.
- [x] `c_summary_subset_label()` should simply call `a_summary_subset()` and add the label.
 Remove `.spl_context` argument and rename `subset_expr` to `filter_expr`.
- [x] Update/add documentation for both functions.

wwojciech:
Fixes #341 

## Checks


munoztd0 (commented):
why did we refactor this ? and why removing the _j I though we agreed on that we kept the suffix ?

wwojciech:
> why did we refactor this ? and why removing the _j I though we agreed on that we kept the suffix ?

The refactoring was done to optimize the code and support the new alternative layout for VS.

I removed `_j` because this is a new function that does not interfere with `tern`, so there is no need to explicitly distinguish it with the `_j` suffix. Besides, we have not yet agreed on a common and consistent naming convention. We already have many functions without the `_j` suffix.
Please, let's keep it as it is for now. Later, if needed, we can do a broader review of the naming conventions globally.

PR #352, opened 2026-05-27
Solves #338 - assertions for safe_t_test

### LINKED ISSUE #338: Add assertions for safe_t_test().
Add assertions for safe_t_test(), specifically for numeric `x` and `y`. We actually should not rely on `stats::t.test()` for this.

wwojciech:
Fixes #338 

## Checks

### DIFF
--- a/R/safe_t_test.r
+++ b/R/safe_t_test.r
@@ -17,6 +17,10 @@
 #' observations) would otherwise stop execution.
 #'
 #' @details
+#' Inputs `x` and `y` must be numeric vectors containing only finite,
+#' non-missing values. In particular, `NA`, `NaN`, `Inf`, and `-Inf` values are
+#' not allowed and will trigger an error before [stats::t.test()] is evaluated.
+#'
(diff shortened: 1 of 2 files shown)


munoztd0:
Looks good to me, but why removing the NA logic and tests, did you talk with @iaugusty about it ?

wwojciech:
> Looks good to me, but why removing the NA logic and tests, did you talk with @iaugusty about it ?

@munoztd0 - You mean the assertions against `NA` in `x` and `y`? That is because: (1) computations on `NA`s do not make sense; and (2) the `safe_t_test()` computations should be safe, i.e. they should not silently remove any `NA`s.

I did not explicitly talk to Ilse about it, but if you look at the code where `safe_t_test()` is used by Ilse, the `NA`s are already removed. See here: https://github.com/johnsonandjohnson/junco/blob/e01232c1d966207696e69bfbad6bb426b313c0f9/R/a_summarize_aval_chg_diff.R#L12-L15

PR #362, opened 2026-05-29
Feature eair: Extra combined statistic required to be able to present "n (EAIR per 100 SY)"

### LINKED ISSUE #361: [Feature Request]: a_eair100_j add extra statistics
### Feature description

Extra combined statistic required to be able to present "n (EAIR per 100 SY)"
Also allow scaling factor other than 100.

iaugusty:
# Pull Request

Fixes #361 

## Checks


tests/testthat/test-patyrs-eair100.R
```diff
@@ -57,6 +57,114 @@ core_lyt <- basic_table(show_colcounts = FALSE, round_type = "sas") |>
...
+  ) * 100
+
+  coeff <- stats::qnorm((1 + conf_level) / 2)
+  lcl <- rdiff - (coeff * se)
+  ucl <- rdiff + (coeff * se)
+
+  eair_diff <- c(rdiff, lcl, ucl)
+}
+# get numbers from single sel_AEDECOD, like dcd A.1.1.1.1, selected treatment group
+eair_numbers <- function(adae, adsl, sel_AEDECOD, comb_group) {
```
munoztd0, line 61: repeating this paragraph isnt ?

DESCRIPTION
```diff
@@ -73,9 +73,11 @@ Suggests:
     pharmaverseadamjnj (>= 0.0.2),
     multcomp,
     lifecycle,
-    dplyr
+    dplyr,
+    ratesci (>= 1.0.0.9000)
```
nikolas-burkoff: I see this is only used for tests now. If it's going to stay like this (and the custom implementation is going to be used) I might actually remove the dependency here and create test cases like below. 

We can then have control over when we update our test cases rather than tests failing because something changed in ratesci (the argument against this is if we think it's likely there are bugs and ratesci and our implementation and ratesci fixes them, then we might miss them)

```r
testthat::test_that("our implementation matches ratesci", {

   our_result <- ...

   # These values have been created by running:
   #  ratesci::package_function(...) using version xxxx of the ratesci package
   ratesci_result <- c("0.0045", "0.33544")

   expect_equal(our_result, ratesci_result)
})
```

DESCRIPTION
```diff
@@ -73,9 +73,11 @@ Suggests:
...
 VignetteBuilder: knitr
 Config/testthat/edition: 3
 Remotes:
     insightsengineering/formatters@main,
-    insightsengineering/tern@main
+    insightsengineering/tern@main,
+    petelaud/ratesci
```
nikolas-burkoff: In terms of keeping our custom implementation v using ratesci - once ratesci is updated on CRAN ( I wouldn't use ratesci until it is), personally I'd prefer to use ratesci rather than maintaining an implementation ourselves as this simplifies thing.

However, adding a dependency on a stats package adds risks in terms of validation. I don't know junco/tern's policy on this so it may be that the control we gain from our own implementation actually outweights the benefits of adding the dependency in our case (especially as the custom implementation is already written)

R/a_eair_j.R
```diff
@@ -233,42 +234,130 @@ NULL
...
+#'   For subjects with an event (`occ_var == "Y"`), this value is converted to years
+#'   as `occ_dy / 365.25` and used as the at-risk exposure time instead of `fup_var`.
+#'   Must be on the same time origin as `fup_var` (i.e., relative to start of exposure).
+#' @param num_p_year (`numeric`)\cr scaling factor for the incidence rate, representing
+#'   the number of person-years used as the denominator base.
+#'   The EAIR is calculated as `num_p_year * n_event / total_person_years`.
+#'   Defaults to `100`, yielding a rate per 100 person-years.
+#'   For example, use `1` for a rate per person-year or `1000` for a rate per
+#'   1000 person-years.
+#' @param count_events (`logical`)\cr if `TRUE`, counts the total number of events
```
nikolas-burkoff: How about `count_multiple_events_per_subj` being `TRUE` or `FALSE` or `count_events_per_subject` being `"first"` (or `"single"`) or `"all"`

R/a_eair_j.R
```diff
@@ -233,42 +234,130 @@ NULL
...
+#'    computed using the method specified by `conf_type`.
+#' }\cr
+#' The list of available statistics (core columns) can also be viewed by
+#' running `junco_get_stats("a_eair100_j")`.
+#' @details
+#' The exposure-adjusted incidence rate (EAIR) is defined as the number of subjects
+#' with at least one occurrence of a specified adverse event divided by the total
+#' at-risk exposure time across all subjects.
+#'
+#' At-risk exposure time is calculated for each subject
```
nikolas-burkoff: I think this needs updating with the new count_events argument

R/a_eair_j.R
```diff
@@ -233,42 +234,130 @@ NULL
...
+#'   `"mn"` (Miettinen-Nurminen score), or `"scas"` (skewness-corrected asymptotic score).
+#'   Defaults to `"wald"`.
 #' @param trt_var (`string`)\cr treatment variable name.
 #' @param ctrl_grp (`string`)\cr control group value.
 #' @param cur_trt_grp (`string`)\cr current treatment group value.
 #' @param inriskdiffcol (`logical`)\cr flag indicating if the function is called within a risk difference column.
-#' @param fup_var (`string`)\cr follow-up variable name.
-#' @param occ_var (`string`)\cr occurrence variable name.
-#' @param occ_dy (`string`)\cr occurrence day variable name.
+#' @param fup_var (`string`)\cr name of the variable containing the total follow-up duration
```
nikolas-burkoff, line 235: These add a lot more detail which I think is great

R/a_eair_j.R
nikolas-burkoff: I also think the default label should change dependent on the value of this, like tern - as it makes it very clear to users what's going on

nikolas-burkoff (commented):
Overall it's looking good. A few comments from me (note I have not gone into detail to check the maths is correct).

I think if both eair is needed by subject and by events having two separate calls is fine - though I would add that as a note in the roxygen comment - it feels unlikely the same output would need both

I think the row labels behaviour makes sense though I don't know if there's an expectation  from others for how it should behave

iaugusty:
@munoztd0 can you take a look into failed checks and pkgdown issues, these seem to be from files not touched in this PR?

iaugusty:
@danielinteractive could you take a look into this PR as well, especially your expertise on derivations of confidence interval for difference in rates - own derivations vs dependency on other statistical package (here ratesci) would be appreciated

current approach is : perform own calculations, in tests confirm results with ratesci package

R/a_eair_diff.R
danielinteractive, line 1: Hi @iaugusty , thanks for reaching out, I had a look at this file in particular. In general I would say it is ok to implement the calculations ourselves and then compare with (hardcoded results from) another package. 
However, I think we should be careful not to have more locations where we put differences of rates code - currently we have https://github.com/insightsengineering/tern/blob/main/R/prop_diff.R and also in our junco package here we have in principle https://github.com/johnsonandjohnson/junco/blob/main/R/estimate_proportion_diff.R. If we could use the `tern` functions as much as possible here that would be good - if the interface from `tern` is not convenient we could also propose a change to that there. If we need something faster then we could also add it here in the `junco` file, but then if possible follow similar conventions for the function names, such that they can also be reused again in other places later.

R/a_eair_diff.R
```diff
@@ -0,0 +1,142 @@
...
+                                  n1,
+                                  x2,
+                                  n2,
+                                  conf_type = c("wald", "waldcc", "mn", "scas"),
+                                  conf_level = 0.95) {
+  conf_type <- match.arg(conf_type)
+  diff_est <- switch(conf_type,
+    waldcc = h_s_eair_diff_wald(x1, n1, x2, n2, conf_level, cc = TRUE),
+    wald   = h_s_eair_diff_wald(x1, n1, x2, n2, conf_level, cc = FALSE),
+    mn     = h_s_eair_diff_mn(x1, n1, x2, n2, conf_level, skew = FALSE),
```
danielinteractive, line 11: I would be relatively optimistic that we have these 3 in `tern` already implemented - did you have a look already?

R/a_eair_diff.R
```diff
@@ -0,0 +1,142 @@
...
+    scas   = h_s_eair_diff_mn(x1, n1, x2, n2, conf_level, skew = TRUE)
```
danielinteractive, line 12: This skewness correction could be interesting and might not be in `tern` yet

R/a_eair_diff.R
iaugusty: @danielinteractive 
incidence rate code is in https://github.com/insightsengineering/tern/blob/v0.9.10/R/incidence_rate.R, rather than proportions.

The design of incidence rate functions, esp how input datasets are utilized, is quite different between `tern` and `junco`. In tern, input dataset requires event and non events in the same input dataset, we have only events in input, non events information is from adsl/alt_counts_df dataset. Our design is easier to generate incidence rates per aedecod, and input dataset requirements is very similar to ADAM ADAE dataset.
As this difference in design exists, it seemed harder to integrate our function within tern.

wrt confidence intervals, in `tern` only confidence intervals for the incidence rates itself are available, not for difference in rates. This would be something we can propose in tern.

danielinteractive: Ah those are incidence rates, ok, I did not see that quickly.

Maybe then for now we could put them in a similar `incidence_rates_diff.R` file in junco and propose an inclusion upstream in tern later. But then we will need more documentation of the calculation functions, including literature references where applicable.

munoztd0 (approved):

munoztd0:
Just FYI, running the scda.test pipeline we get these new snapshot change for tstfae13
image
Is that expected? @iaugusty

DESCRIPTION
```diff
@@ -73,11 +72,14 @@ Suggests:
...
     lifecycle,
-    dplyr
+    dplyr,
+    ratesci (>= 1.1.0)
 VignetteBuilder: knitr
 Config/testthat/edition: 3
 Remotes:
     insightsengineering/formatters@main,
     insightsengineering/tern@main,
+    johnsonandjohnson/pharmaverseadamjnj@dev,
```
nikolas-burkoff, line 83: @munoztd0 do we want to point to dev here?

munoztd0: @nikolas-burkoff well yes we are mirroring dev in all our repos so that junco@main is with pharmaverseadamjnj@main and so on (and our main has been decided to be the stable/CRAN release)

R/utils.R
```diff
@@ -297,3 +297,35 @@ find_missing_chg_after_avisit <- function(df) {
...
+#'
+#' @param x (`list`)\cr with depth of 2.
+#'
+#' @return Transposed version of list where inner elements now are outer elements.
+#' @noRd
+#' @keywords internal
+transpose_named_list <- function(x) {
+  # x: named list of named lists
+  keys <- unique(lapply(x, names))
+  if (!length(keys) == 1) {
```
munoztd0: i believe you meant something like 
```suggestion
 if (length(keys) != 1) {
```

because `!` is applied before `==` in R

munoztd0 (commented):
check my comment on utils but everything else looks good to me!

PR #366, opened 2026-06-01
Investigate package dependencies

### LINKED ISSUE #201: TODO: investigate package dependencies

munoztd0:
# Pull Request

Fixes #201 

@munoztd0
[perf: remove stringi from depencies](https://github.com/johnsonandjohnson/junco/pull/366/commits/bd3d44dfe8272a58cbc6f50bd24be7a6098a6ea5)
@munoztd0
[perf: replace assertthat by checkmate for consistency](https://github.com/johnsonandjohnson/junco/pull/366/commits/4e6aba8d7d015fbc182952342e4454d0382d3064)
@munoztd0
[perf: remove dplyr from package code and use base R](https://github.com/johnsonandjohnson/junco/pull/366/commits/b833fde33b8ccbd31d3a21a22597333d68a1b808)

## Checks


R/pruning_functions.R
```diff
@@ -401,7 +401,7 @@ bspt_pruner <- function(
           # get percent from control group
           pct0 <- cell_values(tt_all_cols, colpath = c(control))[[1]][2]
 
-          if (!dplyr::between(pct0, 0, 1)) {
```
eanokian, line 404: this is good replacement. The problem is that dplyr is used by so many packages that it will be nearly impossible to fully remove its dependency. Examples: vcdExtra, tern and officer import dplyr, therefore we'll need dplyr dependency one way or another. Does it make sense?

munoztd0: CRAN's automated systems (R CMD check --as-cran) issue a NOTE if a package has more than 20 direct hard dependencies (direct hard means "listed in DESCRIPTION" in this case), so even if what you are saying is true , we want to remove as much depedencies as possible from the DESCRITPTION

munoztd0: 1) CRAN's automated systems (R CMD check --as-cran) issue a NOTE if a package has more than 20 direct hard dependencies (direct hard means "listed in DESCRIPTION" in this case)
 
so we want to remove as much depencies as possible from the DESCRITPTION

eanokian (approved):
LGTM.
Question: did you investigate if other dependencies can be removed? are other packages harder?

PR #398, opened 2026-06-26
Add exclude afuns

danielinteractive:
This adds two more `_with_exclude` afuns and the analogue to `leftside()`, namely `rightside()`.

I did not commit my `document()` results because I am using `roxygen2` version 8.0.0 already so there would be lot of unrelated updates...


munoztd0 (approved):
All good to me

wwojciech:
@danielinteractive , @munoztd0 isnt `a_summary_j_with_exclude()` a kind of duplication of `a_summary_subset()` ?

PR #426, opened 2026-08-04
Solves #425: Add factor_by_order(), copy_attributes(), `is_bijection()`.

### LINKED ISSUE #425: [Feature Request]: Add function to create ordered factors from an ordering variable
### Motivation

In some ADaM datasets, character variables often have a corresponding integer variable that defines the ordering of their values.

For example:

```r
TRT01P  <- c("Placebo", "X", "Y")
TRT01PN <- c(1, 2, 3)
```

`TRT01PN` defines the desired order of the levels in `TRT01P`.

Currently, there is a need for a helper function that converts a character vector into a factor (or ordered factor) using a separate ordering vector.

### Proposed functionality

Add a function: `factor_by_order(x, y, ordered = FALSE)`

where:

- `x` is the vector to convert to a factor;
- `y` is an ordering vector defining the level order of `x`;
- `ordered = FALSE` returns a regular factor;
- `ordered = TRUE` returns an ordered factor.

Example:

```r
x <- c("A", "A", "B", "C")
y <- c(3, 3, 2, 1)

factor_by_order(x, y)
# Levels: C B A
```

### Attribute preservation

The function should preserve all attributes of `x` except `levels` and `class`.

### Validation

The function should check that `x` and `y` define a one-to-one correspondence between observed values.

For example, the following should error:

```r
x <- c("A", "B")
y <- c(1, 1)
```

because the ordering is not uniquely defined.

Missing values should be handled explicitly: `NA` in `x` must correspond to `NA` in `y` at the same position; mismatched missingness should result in an error.

wwojciech:
Fixes #425 

## Checks


wwojciech:
Hi @danielinteractive, @gmbecker - would you mind reviewing this PR when you have a chance? Thanks!

R/utils.R
```diff
@@ -329,3 +329,172 @@ transpose_named_list <- function(x) {
...
+#'
+#' Copies attributes from `source` to `target` that are not already present
+#' on `target`. Existing attributes of `target` are preserved.
+#'
+#' @param source An object providing the attributes to copy.
+#' @param target An object to which missing attributes are copied.
+#'
+#' @return The `target` object with any missing attributes copied from `source`.
+#'
+#' @author WW
```
danielinteractive, line 347: I don't think we use `@author` tags in the package, do we?

R/utils.R
```diff
@@ -329,3 +329,172 @@ transpose_named_list <- function(x) {
...
+  n_unique_y <- length(unique(y))
+
+  bijection <- if (n_unique_x != n_unique_y) {
+    FALSE
+  } else {
+    x_first_positions <- match(x, x)
+    all(y == y[x_first_positions], na.rm = TRUE)
+  }
+
+  return(bijection)
```
danielinteractive:
```suggestion
  bijection
```

tests/testthat/test-is_bijection.R
```diff
@@ -0,0 +1,96 @@
+test_that("is_bijection() works as expected for input without missing values", {
+  res <- is_bijection(c("A", "A", "B"), c(1, 1, 2))
+  expect_identical(res, TRUE)
```
danielinteractive: I think there is also an `expect_false()` function which would be nice here

R/utils.R
```diff
@@ -329,3 +329,172 @@ transpose_named_list <- function(x) {
...
+#'
+#' @return A single logical value indicating whether `x` and `y` define a
+#'   bijection.
+#'
+#' @seealso [factor_by_order()]
+#' @keywords internal
+#' @author WW
+#'
+#' @examples
+#' is_bijection(c("A", "A", "B"), c(1, 1, 2))
```
danielinteractive, line 410: can we have examples for non-exported functions? I thought not

danielinteractive (commented):
Thanks @wwojciech , please see comments below

R/utils.R
wwojciech: Actually, I discussed this with @munoztd0, and we agreed that it is beneficial to have it. For instance, in `{rtables}`, Gabe uses his initials. I think this is helpful because it immediately helps identify the authors in case someone from our team has questions about the code or the solutions applied.

R/utils.R
wwojciech: Yes, I checked this with @munoztd0 - we use the `@noRd` tag for non-exported functions.

wwojciech:
> Thanks @wwojciech , please see comments below

Thank you @danielinteractive . Suggestions applied.

R/utils.R
danielinteractive: ah ok - I would just use the git blame functionality for that but ok 👍

danielinteractive (approved):

R/utils.R
```diff
@@ -329,3 +329,172 @@ transpose_named_list <- function(x) {
...
+  checkmate::assert(
+    checkmate::test_character(x) || checkmate::test_numeric(x)
+  )
+  checkmate::assert(
+    checkmate::test_character(y, len = length(x)) ||
+      checkmate::test_numeric(y, len = length(x))
+  )
+
+  # Missingness must align perfectly at the vector level.
+  if (!all(is.na(x) == is.na(y))) {
```
gmbecker: Do you actually need this NA behavior because from a principled perspective it's not correct. Missing means unknown, could be anything, as in it could be a duplicate of one of the preceding values, or a new value. Thus even if there's only one NA per vector, we can't actually determine if its a bijection or not.

R/utils.R
```diff
@@ -329,3 +329,172 @@ transpose_named_list <- function(x) {
...
+  }
+
+  # Check bijection.
+  n_unique_x <- length(unique(x))
+  n_unique_y <- length(unique(y))
+
+  bijection <- if (n_unique_x != n_unique_y) {
+    FALSE
+  } else {
+    x_first_positions <- match(x, x)
```
gmbecker, line 446: this is just going to be 1:n_unique_x if we remove NA support (or leave it in but handle it separately), right?

nevermind I realized you're allowing duplicates as long as the duplication matches between x and y. See newer comment below

gmbecker: Alternatively to this we could just do
```
non_nas_match <- all(tapply(y, x, function(yii) length(unique(yii)) == 1)) # drops NAs
nas_match <- all(is.na(y) == is.na(x)) ## see comment above for issues with this
non_nas_match && nas_match
```

right?

tests/testthat/test-factor_by_order.R
```diff
@@ -0,0 +1,138 @@
...
+  exp <- factor(x, levels = c("B", "A"))
+  expect_identical(res, exp)
+
+  res <- factor_by_order(x, y, ordered = TRUE)
+  exp <- ordered(x, levels = c("B", "A"))
+  expect_identical(res, exp)
+})
+
+test_that("factor_by_order() works as expected for input with all missing values", {
+  x <- NA_character_
```
gmbecker, line 53: Note this generates a factor with no levels, this could very well break things downstream/have consequences. Do we really need to support this case?

```
> factor(NA_character_)
[1] <NA>
Levels: 
> factor(c(NA_character_, NA))
[1] <NA> <NA>
Levels: 
> levels(factor(c(NA_character_, NA)))
character(0)
```

tests/testthat/test-factor_by_order.R
```diff
@@ -0,0 +1,138 @@
...
+  res <- factor_by_order(x, y)
+  expect_identical(res, factor(x))
+
+  res <- factor_by_order(x, y, ordered = TRUE)
+  expect_identical(res, ordered(x))
+})
+
+test_that("factor_by_order() works as expected for empty vectors", {
+  x <- character(0)
+  y <- integer(0)
```
gmbecker, line 65: This is another case I'm not sure we need to support. This behavior is technically consistent but I'm not sure how safe it is when plugged into a larger workflow.

Specifically, having a factor with no observations is fine, having a character with no observations is fine, I'm not sure about a factor with no levels though; its not really meaningfully a factor at that point

tests/testthat/test-factor_by_order.R
```diff
@@ -0,0 +1,138 @@
...
+  exp <- ordered(x, levels = c("B", "A"))
+  expect_identical(res, exp)
+})
+
+test_that("factor_by_order() works as expected for factors with unobserved levels", {
+  x <- factor(c("A", "A", "B"), levels = c("A", "B", "C"))
+  y <- c(2, 2, 1)
+
+  res <- factor_by_order(x, y)
+  exp <- factor(x, levels = c("B", "A"))
```
gmbecker, line 92: Is dropping the unobserved levels really the correct behavior? Whats the use-case justification here?

gmbecker (commented):
overall looks reasonable and well documented/tested. I have some questions about the behaviors for some of the corner cases, see comments inline

R/utils.R
wwojciech: I agree. I added this because `is_bijection()` is a helper for `factor_by_order()`, and I wanted to ensure that two vectors with aligned NAs could still be converted to factors without any issues.

That said, I think `is_bijection()` should either:
- not accept any NAs at all (from the principled perspective you mentioned), or
- simply ignore/remove any NAs before checking for a bijection.

What do you think, @gmbecker?

R/utils.R
wwojciech: Not quite; even without any `NA`s, this only checks one direction of the injection. For example,
```r
x <- c("A", "B")
y <- c(1, 1)
```
is clearly not an injection (and therefore not a bijection), but the `tapply()` approach would incorrectly identify it as one.
So we'd also need to check the other direction too; at that point, I think the code becomes more complex than my original solution.

tests/testthat/test-factor_by_order.R
wwojciech: I wanted `factor_by_order()` to behave as similarly as possible to `factor()`. Since `factor()` accepts inputs containing only `NA`s, I wanted `factor_by_order()` to handle this case in the same way.

Essentially, I see `factor_by_order()` as being more transparent, and it is the user's responsibility to verify whether the input is appropriate.

tests/testthat/test-factor_by_order.R
wwojciech: Similar to the above, I wanted `factor_by_order()` to behave as similarly as possible to `factor()`. Since `factor()` accepts inputs containing empty strings, I wanted `factor_by_order()` to handle this case in the same way.

tests/testthat/test-factor_by_order.R
wwojciech: I thought about the use case: we already have a factor variable `x` (say, in a subset of a dataset where some levels are not observed), and we want to reorder the levels according to `y`.

Yeah, I agree that dropping levels is risky here. I think the safest solutions would be for the function to:
- fail,
- return the unchanged vector (with a warning?), or,
- reorder the levels while keeping the positions of the unobserved levels unchanged and only reordering the observed levels. 

However, none of these options is perfect.....

wwojciech:
> overall looks reasonable and well documented/tested. I have some questions about the behaviors for some of the corner cases, see comments inline

Thank you @gmbecker for valuable comments. See my responses above.

R/utils.R
gmbecker: ah yes, you're right of course, I wrote that too quickly. I think it is sufficient under the constraint that `n_unique_x == n_unique_y` and when operating on only the non-NAs (na matching is easy to do via `is.na`), which I was applying in my head when I wrote that.

R/utils.R
gmbecker: I mean we only need this for the one use case, but from a purely design perspective under the assumption one might want this function elsewhere, I'd add an argument `match_nas` which defaults to `FALSE` and then set it to `TRUE` when we call it from `factor_by_order`. If its `FALSE` I'd throw an error, and if its `TRUE` have the behavior you have now where they have to positionally match

tests/testthat/test-factor_by_order.R
gmbecker: This is reasonable but should be called out explicitly in the docs if it isn't already

tests/testthat/test-factor_by_order.R
gmbecker: I just got off a call with @jkovach2 and we decided that for this function the best approach is to assume the user does not want to retain any unobserved levels, and document that explicitly.

If a user wants to shuffle the order of a factor and does want (at least some) unobserved levels retained they will need to write code to do that and should not use `factor_by_order`

tests/testthat/test-factor_by_order.R
gmbecker: as above, please explicitly document this to be safe

gmbecker (changes requested):
Please update the documentation as indicated in my comments, other than that looks good

tests/testthat/test-factor_by_order.R
wwojciech: I have updated the `@return` documentation block for `factor_by_order()` to explicitly call this out.

tests/testthat/test-is_bijection.R
wwojciech: Hey @danielinteractive,

Just realised, the following holds:
```
expect_identical(x, TRUE) => expect_true(x)
```
but the converse, may not hold. 

Example:
```r
x <- c(res = TRUE)

expect_true(x) # passes
expect_identical(x, TRUE) # fails
Error: Expected `x` to be identical to TRUE.
Differences:
`names(actual)` is a character vector ('res')
`names(expected)` is absent
```
Of course, the behavioral difference is negligible in practice, but still... 
What do you think?

gmbecker (approved):
please resolve:

```
* checking Rd line widths ... NOTE
  Rd file 'factor_by_order.Rd':
    \examples lines wider than 100 characters:
       # Error: `x` and `y` must define a bijection between their unique non-NA values; NA values must correspond.
```

Other than that looks good to go

PR #437, opened 2026-08-18
436 fr function format sigfig j

### LINKED ISSUE #436: [Feature Request]: formatting function to adhere to PK specifications
### Feature description

PK specifications to display descriptive statistics 
- up to 1000 with 3 significant figures
- values >= 1000, the whole integer is to be reported

Image

iaugusty:
# Pull Request

Fixes #436

## Checks


R/jjcsformats.R
```diff
@@ -368,3 +368,110 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
+#' fmt2(c(1234.5, 0.00567))
+#'
+#'
+#' format_sigfig_j(3, trail_zero = TRUE)(0.01)
+#' format_sigfig_j(3, trail_zero = FALSE)(0.01)
+format_sigfig_j <- function(
+  sigfig,
+  format = "xx",
+  whole_integer = TRUE,
+  zero_threshold = 10^(-2 * sigfig),
```
munoztd0: why is this -2 ??

I believe it should be 10^(sigfig-1)

100 for sigfig 3
1000 for sigfig 4

R/jjcsformats.R
```diff
@@ -368,3 +368,113 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
+#' @param digits (`integer(1)`)\cr number of significant figures to display.
+#' @param round_type (`character(1)`)\cr rounding method. See [formatters::format_value()] for details.
+#' @param whole_integer (`logical(1)`)\cr if `TRUE`, preserves the full integer part when
+#'   `x` has more integer digits than `digits` (i.e., scale is clamped to 1).
+#' @param zero_threshold (`numeric(1)`)\cr values with `abs(x) < zero_threshold` are snapped
+#' to zero before rounding. Defaults to `0` (no snapping).
+
+#' @return numeric vector of the same length as `x`, rounded to `digits` significant figures.
+#' @keywords internal
+signif_j <- function(x, digits = 6, round_type = valid_round_type, whole_integer = FALSE, zero_threshold = 0) {
```
wwojciech: I'd aim for the most straightforward extension of `base::signif()`: preserve its behavior while allowing custom rounding, including the support for `round_fun` that does not support negative digits; and an optional `zero_threshold`.

I propose:

```r
signif_j <- function(
  x,
  digits = 6,
  zero_threshold = 0,
  round_fun = round_fmt,
  ...
)
```
Here, `...` are passed as arguments to `round_fun`.

R/jjcsformats.R
```diff
@@ -368,3 +368,113 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
+
+  ok <- is.finite(x) & x != 0
+  # see R documentation signif is the same as rounding with  dig - ceiling(log10(abs(x)) digits
+  exponent <- ceiling(log10(abs(x[ok])))
+  new_round <- digits - exponent
+  if (whole_integer) {
+    # to ensure entire integer part is presented if x is greater than 10^digits
+    round_integer <- new_round < 0
+    new_round[round_integer] <- 0
+  }
```
wwojciech: Could we perhaps remove `whole_integer` here? What is the use case for it? As I mentioned above, I think it would be good to keep this function as consistent as possible with `base::signif()`.

R/jjcsformats.R
```diff
@@ -368,3 +368,113 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
+  function(x, round_type = valid_round_type, ...) {
+    if (!is.numeric(x)) stop("`format_sigfig_j` cannot be used for non-numeric values. Please choose another format.")
+    x_f <- signif_j(
+      x,
+      digits = sigfig,
+      round_type = round_type,
+      whole_integer = whole_integer,
+      zero_threshold = zero_threshold
+    )
+    num <- formatC(x_f, digits = sigfig, format = "fg", flag = flag)
```
wwojciech: This is easier to read and makes the PK reporting rule explicit. I also think it is safer because the threshold logic is handled directly here rather than being embedded in `signif_j()`:
```r
  # Vectorized conditional handling
  ifelse(
    x < threshold,
    formatC(signif_j(x, digits = sigfig, round_fun = round_fun, ...), digits = sigfig, flag = flag), # Up to threshold: round to sigfig significant digits
    formatC(round_fun(x, digits = 0, ...)) # At or above threshold: report the whole integer
  )
```
(The details of the arguments can of course be adjusted.)

wwojciech (changes requested):
Overall, the idea looks good to me. Nevertheless, it would be good to clarify and potentially improve some of the design assumptions and implementation details, as indicated in my comments.

Please also check whether the CI/CD failure is related to the new code.

munoztd0:
> Overall, the idea looks good to me. Nevertheless, it would be good to clarify and potentially improve some of the design assumptions and implementation details, as indicated in my comments.
> 
> Please also check whether the CI/CD failure is related to the new code.

CI/CD issues are not related to us, just ubuntu stuff, will resolveshortly

munoztd0 (approved):

R/jjcsformats.R
gmbecker: I'm not sure I agree. I would suggest changing the name of the function (I agree having something that is named signif_j should probably be a drop in replacement for signif, but that doesn't do everything we need, as illustrated by ilse's code. 

I'd suggest something like
```
roundSAS <- function(x, digits) round_fmt(x, digits, type = "sas")
#' @param trailing_period Should a trailing `.` be rendered when x rounds to a whole number. Defaults to `FALSE`
modified_sigfig_j <- function(x, 
                              digits,
                              round_fun = roundSAS,
                              whole_integer = TRUE,
                              zero_threshold = 0, 
                              trailing_period = FALSE) {
...
```

so that we have a function which does exactly what we want without confusion. I'm not married to the name, of course, the point is to have it not just be `signif_j` to alleviate the types of concern wojtek has

wwojciech (changes requested):
A few areas should be strengthened:
1. Add argument assertions for `modified_signif_j()` (specifically `zero_threshold` should be very well-assserted).
2. Improve documentation for `modified_signif_j()`:
 - describe in detail how this function differs from `base::signif()`.
 - add illustrative examples
 - use lifecycle badge.
3. Prepare dedicated unit test for various values of the arguments.

Please use `mantel_fleiss_crit()` ( https://github.com/pharmaverse/tern/pull/1513 ) as a reference.

R/jjcsformats.R
```diff
@@ -384,6 +390,8 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
 modified_signif_j <- function(x, digits = 6, round_type = valid_round_type, whole_integer = FALSE, zero_threshold = 0) {
   stopifnot(length(digits) == 1, is.numeric(digits))
```
wwojciech: Use checkmate instead.

R/jjcsformats.R
```diff
@@ -384,6 +390,8 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
 
+  checkmate::assert_numeric(zero_threshold, lower = 0, upper = 10^(-digits), len = 1)
```
wwojciech: is missing value allowed? I guess not - use `assert_number()` instead.
How about other args, why are they not asserted?

tests/testthat/test-jjcsformats.R
```diff
@@ -443,3 +443,70 @@ test_that("format_sigfig_j used in rtables framework as format w/wout trailing z
   rslt
   expect_snapshot(cran = TRUE, rslt)
 })
+
+test_that("explicit modified_signif_j tests", {
```
wwojciech: Tests should be specific: 1 test = 1 issue (or a few very similar issues). Please avoid putting all the checks into 1 tests. Each test should focus on a specific functionality or scenario so that it is clear what exactly is being verified.

R/jjcsformats.R
wwojciech, line 349: why is `str` not asserted?

R/jjcsformats.R
```diff
@@ -370,12 +370,18 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
 }
 
 #' @noRd
-#'
+#' @description `r lifecycle::badge('stable')`
+#' Modified version of base function `signif` on 3 topics.
```
wwojciech: Modified version of [base::signif()] with three extensions:

R/jjcsformats.R
```diff
@@ -370,12 +370,18 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
+#' Rounding approach: also non-iec rounding can be used (`round_type`), through function `formatters::round_fmt`.\cr
```
wwojciech: Rounding is performed using [formatters::round_fmt()].

R/jjcsformats.R
```diff
@@ -370,12 +370,18 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
+#' Very small absolute values can be considered zero (`zero_threshold`).\cr
+#' Showing all significant figures to the left of the decimal marker (`whole_integer`)
+#' rather than the specified number of significant figures.
```
wwojciech: #' Very small absolute values can be treated as zero (`zero_threshold)`.
#' All significant figures to the left of the decimal marker can be shown (`whole_integer`), rather than limiting the result to the specified number of significant figures.

Please use roxygen list.

R/jjcsformats.R
```diff
@@ -370,12 +370,18 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
 #' @param x (`numeric`)\cr numeric vector to round.
```
wwojciech: Should start with upper case. 
I think there is not need to repeat "numeric vector" - this is already clear from (`numeric`)\cr. 
"Numbers to round." - should be ok.

R/jjcsformats.R
```diff
@@ -370,12 +370,18 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
 #' @param digits (`integer(1)`)\cr number of significant figures to display.
 #' @param round_type (`character(1)`)\cr rounding method. See [formatters::format_value()] for details.
```
wwojciech: Add info that this arg is passed directly to [formatters::round_fmt()]

R/jjcsformats.R
wwojciech: Why weren’t the examples added?

R/jjcsformats.R
```diff
@@ -384,6 +390,8 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
 modified_signif_j <- function(x, digits = 6, round_type = valid_round_type, whole_integer = FALSE, zero_threshold = 0) {
```
wwojciech: I think Gabe and I suggested to use `round_fun` arg to allow for a custom rounding. Why was this not implemented?

wwojciech (changes requested):
Changes required.

R/jjcsformats.R
iaugusty: the current implementation was easier for the setup of the formatting function, which needs round_type as argument. In the other design, an extra conversion from round_fun to round_type was needed. As round_type is used throughout junco, I believe this approach is not that awkward.

R/jjcsformats.R
iaugusty: function was not described in helpfiles, now added this function in help file for format_sigfig_j

R/jjcsformats.R
```diff
@@ -368,3 +368,134 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
+#'   round_type = "sas"
+#' )
+#' @keywords internal
+modified_signif_j <- function(x, digits = 6, round_type = valid_round_type, whole_integer = FALSE, zero_threshold = 0) {
+  round_type <- match.arg(round_type)
+
+  checkmate::assert_numeric(x)
+  checkmate::assert_integerish(digits, lower = 0, len = 1, any.missing = FALSE)
+  checkmate::assert_number(zero_threshold, lower = 0, upper = 10^(-digits))
+  checkmate::assert_logical(whole_integer, len = 1, any.missing = FALSE)
```
wwojciech: checkmate::assert_flag()

R/jjcsformats.R
```diff
@@ -368,3 +368,134 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
+#'   zero_threshold = 1e-6,
+#'   whole_integer = TRUE,
+#'   round_type = "sas"
+#' )
+#' @keywords internal
+modified_signif_j <- function(x, digits = 6, round_type = valid_round_type, whole_integer = FALSE, zero_threshold = 0) {
+  round_type <- match.arg(round_type)
+
+  checkmate::assert_numeric(x)
+  checkmate::assert_integerish(digits, lower = 0, len = 1, any.missing = FALSE)
```
wwojciech: checkmate::assert_int()

R/jjcsformats.R
```diff
@@ -368,3 +368,134 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
     paste0("(", res[1], ", ", res[2], ")")
   }
 }
+
+#' @title Format numeric values by significant figures.
+#' @description `r lifecycle::badge('experimental')`\cr
```
wwojciech: Add 1 line space here.

R/jjcsformats.R
```diff
@@ -368,3 +368,134 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
+#' Format numeric values by significant figures with controllable round_type,
+#' whole_integer and trailing zeros.\cr
+#' The underlying function for significant figures is `modified_signif_j()`, which is a modified version for
+#' [base::signif()]
+#' @details `format_sigfig_j()`\cr A function factory that produces formatting functions to round values to a
```
wwojciech: why `format_sigfig_j()`\cr ?

This is not the way to distinguish between different functions in roxygen. Please see `a_summary_subset.R` how this can be achieved in a proper way.

R/jjcsformats.R
```diff
@@ -368,3 +368,134 @@ jjcsformat_range_fct <- function(str, censor_char = "+") {
...
+format_sigfig_j <- function(
+  sigfig,
+  format = "xx",
+  whole_integer = TRUE,
+  zero_threshold = 10^(-2 * sigfig),
+  drop0trailing = FALSE
+) {
+  checkmate::assert_integerish(sigfig)
+  format <- gsub("xx\\.|xx\\.x+", "xx", format)
+  checkmate::assert_choice(format, c("xx", "xx / xx", "(xx, xx)", "xx - xx", "xx (xx)", "xx, xx"))
```
wwojciech, line 436: format can also by a function?, so better not to fix choices here. 
Just check if this is a function or a string.

wwojciech (changes requested):

R/jjcsformats.R
iaugusty: as discussed, format here should be a string, not a formatting function

wwojciech (approved):
As discussed today with @iaugusty - we clarified all the issue, and once solved - I am happy to approve it. Thank you @iaugusty .

PR #455, opened 2026-09-01
Solves #454: Add `get_complete_cases()`.

### LINKED ISSUE #454: [Feature Request]: Add `get_complete_cases()` helper function.
### Description

Add a helper function, `get_complete_cases()`, to remove observations containing missing values from a data frame.

### The function should:

1. Return only complete observations while preserving the original data-frame structure.
2. Return the original data frame unchanged when there are no missing values.
3. Issue a warning when observations containing missing values are removed.
4. Include the number of omitted observations in the warning.

wwojciech:
Fixes #454 

## Checks


R/utils.R
```diff
@@ -579,3 +579,55 @@ strict_match <- function(x, y, odd = TRUE) {
...
+#' @export
+#' @author WW
+#'
+#' @examples
+#' df <- data.frame(a = c(1:5, NA), b = c(NA, letters[1:5]))
+#' df
+#'
+#' get_complete_cases(df)
+#'
+get_complete_cases <- function(df, additional_message = ".") {
```
danielinteractive: maybe good to add another argument that can be used to omit the warning, e.g. `quiet = FALSE` as default. Because I can imagine that we often don't want the warning when this runs inside a function called by `rtables`.

R/utils.R
```diff
@@ -579,3 +579,55 @@ strict_match <- function(x, y, odd = TRUE) {
...
+  checkmate::assert_data_frame(df)
+  checkmate::assert_string(additional_message)
+
+  is_complete <- complete.cases(df)
```
danielinteractive:
```suggestion
  is_complete <- stats::complete.cases(df)
```

R/utils.R
wwojciech: Sure, I was even thinking about this for a moment :)

danielinteractive (approved):

wwojciech:
Thank you very much @danielinteractive for reviewing this PR.

PR #463, opened 2026-09-05
Add conditional proportion estimation function

danielinteractive:
Adds `a_cond_proportion_j` and required stats and description function, with tests.

@wwojciech fyi this is what we discussed yesterday


R/cond_proportion_j.R
```diff
@@ -0,0 +1,213 @@
...
+  checkmate::assert_int(denom_limit, lower = 0)
+
+  rsp <- if (checkmate::test_atomic_vector(df)) {
+    as.logical(df)
+  } else {
+    as.logical(df[[.var]])
+  }
+
+  n_obs <- length(rsp)
+  n_rsp <- sum(rsp)
```
wwojciech, line 101: Do we allow `NA`s in `rsp`?

I guess this is about the "observed" response so,
```r
n_rsp <- sum(rsp, na.rm = TRUE)
n_obs <- sum(!is.na(rsp))
```

or maybe better to introduce `na.rm` argument or simply assert that no `NA`s are allowed.

R/s_cond_proportion_j.R
```diff
@@ -0,0 +1,213 @@
...
+    switch(
+      n = n_obs,
+      N_row = .N_row,
+      N_col = .N_col
+    )
+
+  p_hat <- ifelse(denom_val > 0, n_rsp / denom_val, 0)
+
+  # Adaptive method selection based on observed data and limits.
+  use_exact <- (n_obs < denom_limit) || (n_rsp <= num_limit) || (n_rsp >= (n_obs - num_limit))
```
wwojciech: `n_rsp`, `n_obs` are based on `rsp`, but the actual denominator can be .`N_row` or `.N_col`. Hence, the logic for choosing "exact" vs "wald" may not correspond to the denominator used to calculate the CI.

Maybe better to use: 
```r
use_exact <- (denom_val < denom_limit) ||
  (n_rsp <= num_limit) ||
  (n_rsp >= (denom_val - num_limit))
```

R/cond_proportion_j.R
```diff
@@ -0,0 +1,213 @@
...
+
+  # Adaptive method selection based on observed data and limits.
+  use_exact <- (n_obs < denom_limit) || (n_rsp <= num_limit) || (n_rsp >= (n_obs - num_limit))
+  method <- if (use_exact) "clopper-pearson" else "wald"
+
+  prop_ci <- switch(
+    method,
+    "clopper-pearson" = prop_clopper_pearson(rsp, n = denom_val, conf_level),
+    "wald" = prop_wald(rsp, n = denom_val, conf_level)
+  )
```
wwojciech, line 125: How about allowing user to chose the method by providing 2 arguments: `method1`, `method2` ?

R/cond_proportion_j.R
```diff
@@ -0,0 +1,213 @@
...
+  list(
+    "n_prop" = formatters::with_label(c(n_rsp, p_hat), "Responders"),
+    "prop_ci" = formatters::with_label(
+      x = 100 * prop_ci,
+      label = d_cond_proportion_j(
+        conf_level,
+        long = long,
+        num_limit = num_limit,
+        denom_limit = denom_limit
+      )
```
wwojciech, line 136: Probably better to compute separately above (mainly to increase the code readability).

R/cond_proportion_j.R
```diff
@@ -0,0 +1,213 @@
...
+    paste0(
+      "Wald if n >= ",
+      denom_limit,
+      " and x > ",
+      num_limit,
+      ", else Clopper-Pearson"
+    )
+  } else {
+    "Wald / Clopper-Pearson"
+  }
```
wwojciech, line 165: Cannot we identify "Wald" or "Clopper-Pearson"? Currently I am getting:
```r
> d_cond_proportion_j(
     0.95,
     long = FALSE,
     num_limit = 10,
     denom_limit = 10
 )
[1] "95% CI (Wald / Clopper-Pearson)"
```
and I think it would be good to know which one of these 2 methods was used.

R/cond_proportion_j.R
```diff
@@ -0,0 +1,213 @@
...
+#'
+#' @export
+s_cond_proportion_j <- function(
+  df,
+  .var,
+  conf_level = 0.95,
+  long = FALSE,
+  num_limit = 0,
+  denom_limit = 10,
+  denom = c("n", "N_col", "N_row"),
```
wwojciech, line 73: Missing assertion for `denom`.

R/cond_proportion_j.R
```diff
@@ -0,0 +1,213 @@
...
+      n = n_obs,
+      N_row = .N_row,
+      N_col = .N_col
+    )
+
+  p_hat <- ifelse(denom_val > 0, n_rsp / denom_val, 0)
+
+  # Adaptive method selection based on observed data and limits.
+  use_exact <- (n_obs < denom_limit) || (n_rsp <= num_limit) || (n_rsp >= (n_obs - num_limit))
+  method <- if (use_exact) "clopper-pearson" else "wald"
```
wwojciech, line 119: maybe `ifelse()`?

R/s_cond_proportion_j.R
```diff
@@ -0,0 +1,213 @@
...
+) {
+  checkmate::assert_flag(long)
+  tern::assert_proportion_value(conf_level)
+  checkmate::assert_int(num_limit, lower = 0)
+  checkmate::assert_int(denom_limit, lower = 0)
+
+  rsp <- if (checkmate::test_atomic_vector(df)) {
+    as.logical(df)
+  } else {
+    as.logical(df[[.var]])
```
wwojciech: would be nice to assert if column `.var` is in `df`.
Moreover: make sure that `df`/`df[[.var]]` can be nicely converted to logical, because it may happen that, e.g.: 
```r
as.logical(c("Y", "N"))
[1] NA NA
```

R/cond_proportion_j.R
```diff
@@ -0,0 +1,213 @@
...
+  n_obs <- length(rsp)
+  n_rsp <- sum(rsp)
+
+  denom_val <- match.arg(denom) |>
+    switch(
+      n = n_obs,
+      N_row = .N_row,
+      N_col = .N_col
+    )
+
```
wwojciech, line 109: I would validate `n_rsp <= denom_val` ( I guess this should not happen, but just in case of some wrong computations by the rtables - so that we do not end up with with an invalid proportion (> 1).

wwojciech (changes requested):
Thank you @danielinteractive - just commented on view issues that I spotted.

munoztd0 (changes requested):
lints issues

R/cond_proportion_j.R
danielinteractive: we can add that later if we really need it. until then the YAGNI law applies 😄

R/s_cond_proportion_j.R
danielinteractive: good point thanks

R/cond_proportion_j.R
danielinteractive: not needed here (that is only for vectorization)

R/cond_proportion_j.R
danielinteractive: hm I think it is well readable, also with switch we don't need to compute something we won't use

R/cond_proportion_j.R
wwojciech: For a moment I though I may need it to my Mantel-Fleiss criterion, but later realised, that of course I do not :)

R/cond_proportion_j.R
danielinteractive: no we cannot because this label needs to be the same across columns, where then different methods might have been chosen. This would only work with some kind of footnote symbols etc. directly at the values

danielinteractive: probably also here YAGNI, but checking it with team

wwojciech: @danielinteractive - as a side note - maybe better to have 1 arg: `limit`, which is a numeric vec. of length 2, instead of 2 args? not sure.

danielinteractive: no... much better readable like this because the limit is not lower/upper on the same quantity here.

danielinteractive: ok confirmed they don't need symbols

munoztd0 (changes requested):
could not find function "s_cond_proportion_j"

Thats the issue with no exported functions in examples either wrap it in dontrun{} pr remove it

wwojciech:
> could not find function "s_cond_proportion_j"
> 
> Thats the issue with no exported functions in examples either wrap it in dontrun{} pr remove it

I guess we agreed for `@noRd`

munoztd0:
> > could not find function "s_cond_proportion_j"
> > Thats the issue with no exported functions in examples either wrap it in dontrun{} pr remove it
> 
> I guess we agreed for `@noRd`

Yes indeed

danielinteractive:
Ah yeah no problem I just had forgotten to regenerate the documentation incl. the namespace file. These functions should be exported indeed.

danielinteractive:
@munoztd0 maybe you can help me with this unrelated (?) snapshot update. I get different numbers locally.

munoztd0 (approved):

wwojciech:
@munoztd0 - please do not merge it yet. After discussion with `{tern}` authors - this function will go to tern - as it better fits there. I will soon make all the necessary updates and close this PR.

danielinteractive:
@wwojciech I sent you a message

wwojciech:
Hi @munoztd0 - I’m very sorry for the confusion caused by my message asking you to hold off on merging this PR. I made a mistake - I was thinking about one of my other PRs, had way too many windows open, and accidentally posted the message in this PR.

I sincerely apologise for the confusion. Please go ahead and merge this PR.

Thank you, @danielinteractive, for pointing this out, and sorry everyone for the confusion!

Issue #453, opened 2026-08-31
[Bug]: `s_freq_j()` returns empty count lists instead of 0.


wwojciech:
### What happened?

The `s_freq_j()` function appears to return empty lists for its count-related outputs when there are no matching responses. In this case, `val = TRUE`, but all values of `rsp` are `FALSE`, so the expected count should be `0`.

```r
df <- data.frame(id = 1:2, rsp = c(FALSE, FALSE))

x_stat <- junco:::s_freq_j(
  df = df,
  .var = "rsp",
  val = TRUE,
  .df_row = NULL,
  alt_df = NULL,
  parent_df = NULL,
  id = "id",
  denom = "n_df",
  countsource = "df"
)
```

Current result

```
x_stat
$count
list()

$count_unique
list()

$count_unique_fraction
list()

$count_unique_denom_fraction
list()
```

wwojciech:
Hi @iaugusty - can you please have a look into this issue? I need to use this function for my table. Thank you in advance!

iaugusty:
Hi @wwojciech, please convert the input variable rsp to a factor

```
df <- data.frame(id = 1:2, rsp = c(FALSE, FALSE))
df$rsp <- factor(df$rsp, levels = c(TRUE, FALSE))
x_stat <- junco:::s_freq_j(
  df = df,
  .var = "rsp",
  val = TRUE,
  .df_row = NULL,
  alt_df = NULL,
  parent_df = NULL,
  id = "id",
  denom = "n_df",
  countsource = "df"
)
```

Updated result:

```
$n_altdf
n_altdf 
      0 

$n_df
n_df 
   2 

$n_rowdf
n_rowdf 
      0 

$n_parentdf
n_parentdf 
         2 

$denom
denom 
    2 

$count
$count$`TRUE`
count 
    0 


$count_unique
$count_unique$`TRUE`
count_unique 
           0 


$count_unique_fraction
$count_unique_fraction$`TRUE`
count_unique            p 
           0            0 


$count_unique_denom_fraction
$count_unique_denom_fraction$`TRUE`
count_unique            d            p 
           0            2            0 
```

wwojciech:
Thank you, @iaugusty.

It works well after converting to `factor`.

Questions related to your answer:

1. Is there any technical reason why this must be a `factor`?
2. If there is such a reason, please add an assertion that `.var` must be a `factor`.
[...]

PR #469, opened 2026-09-09
Fix: Added factor assertion on `.var` in `s_freq_j`

munoztd0:
# Pull Request

Fixes #453 by adding factor assertion on .var in s_freq_j

## Checks


R/a_freq_j.R
```diff
@@ -4,6 +4,8 @@
 #' and (optional) relative risk columns
 #'
 #' @inheritParams proposal_argument_convention
+#' @param .var (`string`) Name of a categorical analysis variable in `df`.
+#' The variable must be a `factor`, `character`, or `logical` vector.
```
wwojciech: #' @param .var (`characgter(1)`) \cr
#'  Name of a categorical analysis variable in `df`. 
#'  Must be a `factor`, `character`, or `logical` column.

R/a_freq_j.R
```diff
@@ -94,12 +96,43 @@ s_freq_j <- function(
...
+    classes = c("factor", "character", "logical")
+  )
+
+  if (is.logical(df[[.var]])) {
+    df[[.var]] <- factor(df[[.var]], levels = c(TRUE, FALSE))
+  } else if (is.character(df[[.var]])) {
+    var_levels <- unique(c(as.character(val), df[[.var]]))
+    var_levels <- var_levels[!is.na(var_levels)]
+    df[[.var]] <- factor(df[[.var]], levels = var_levels)
+  }
```
wwojciech: If converting to factor is really needed, then:
```r
var_levels <- if (is.logical(df[[.var]])) {
  c(TRUE, FALSE)
} else if (is.character(df[[.var]])) {
  unique(c(val, na.omit(df[[.var]])))
}
df[[.var]] <- factor(df[[.var]], levels = var_levels)
```
It should be already asserted somewhere above that `val` is compatible already with the type of `df[[.var]]`.

R/a_freq_j.R
```diff
@@ -94,12 +96,43 @@ s_freq_j <- function(
...
+
+  if (!is.null(.df_row) && .var %in% names(.df_row)) {
```
wwojciech: what is `.var %in% names(.df_row)` is `FALSE`? Then the code below will fail....

R/a_freq_j.R
```diff
@@ -4,6 +4,8 @@
 #' and (optional) relative risk columns
 #'
 #' @inheritParams proposal_argument_convention
+#' @param .var (`string`) Name of a categorical analysis variable in `df`.
+#' The variable must be a `factor`, `character`, or `logical` vector.
 #' @param val (`character` or NULL)\cr
```
wwojciech: cannot be `val = TRUE` ? Then, this is not a character. Also it should be a scalar, so `character(1)`, `logical(1)`, ......

wwojciech, line 9: `val` should be properly asserted.

wwojciech (changes requested):
Few changes + I do not understand the reason why `df[[.var]]`, `.df_row[[.var]]` must be converted to factor. Lets have a quick call about it.

R/a_freq_j.R
```diff
@@ -4,6 +4,8 @@
 #' and (optional) relative risk columns
 #'
 #' @inheritParams proposal_argument_convention
+#' @param .var (`string`) Name of a categorical analysis variable in `df`.
```
wwojciech: (`character(1)`) \cr

R/a_freq_j.R
```diff
@@ -103,6 +106,9 @@ s_freq_j <- function(
     df <- alt_df
   }
 
+  checkmate::assert_names(names(df), must.include = .var)
+  checkmate::assert_class(df[[.var]], classes = c("factor"))
```
wwojciech: `c()` is not needed :)

munoztd0: yeah that was ultra lazy me..
```suggestion
  checkmate::assert_class(df[[.var]], classes = "factor")
```

wwojciech (approved):
thank you @munoztd0 just please check if this does not cause any issues in upstrea, dep. - I mean, maybe good to run scda?

munoztd0:
all good -> https://github.com/insightsengineering/scda.test/pull/254

> thank you @munoztd0 just please check if this does not cause any issues in upstrea, dep. - I mean, maybe good to run scda?

